PR #21, opened 2019-11-12
Finshed two problems using DFS and Set.

Shridhar-2205:
I used DFS here, because I used BFS in interview. I was not able to solve the Sudoku in interview. Do we need to do BFS/DFS in Sudoku problem or SET is fine.


sehaj-grover:
Set is fine

PR #45, opened 2020-04-20
19th April 2020

vaidyapawan:
36. Valid Sudoku
515. Find Largest Value in Each Tree Row


devashishnyati:
Good job Pawan!
The time and space complexity for valid sudoku can be said constant.

PR #47, opened 2020-04-21
Completed Competitive coding-5: Please review

KBangera:


Valid_Sudoku.java
```diff
@@ -0,0 +1,58 @@
+// Time Complexity :O(n)
```
devashishnyati, line 1: Time can also be said as constant.

devashishnyati:
Good job Karthik!

PR #49, opened 2020-05-25
[Competitive-Coding-5] Please review

yosnehal:


36_ValidSudoku.java
```diff
@@ -0,0 +1,31 @@
+// Time Complexity : O(1)
```
iivaishnavii, line 1: We are going through all elements.

ssgandham: yes, but at max we'll got through 81 elements right which is constant

iivaishnavii: Ah, I overlooked that. You are right!

PR #51, opened 2020-05-25
solutions to all problems, please review

domzhaomathematics:


valid_sudoku.py
```diff
@@ -0,0 +1,40 @@
+
+#Problem 2: Valid Sudoku
+#Time Complexity O(n) , size of sudoku, which is fixed
+#Space Complexity O(n), size of sudoku, which is fixed
+
+'''
+Brute force: Keep 27 hashsets (constant space, since always 9x9)
+one for each row and one for each column, one for each block. Traverse the hashset accordingly
+and check if element is not in hashset and smaller than 9 , we continue, else , return false.
+Can we do it with less space?
```
chiragshahi, line 10: We can create only 1 set and store values other than '.' in it along with its row/col/block. if at any time, that string is already in the set we return false
See my code in Java:
        Set seen = new HashSet<String>();
        for (int row = 0; row < 9; row++) {
            for (int col = 0; col < 9; col++) {
                char number = board[row][col];
                if (number != '.') {
                    if (!seen.add(number + "R" + row) ||
                       (!seen.add(number + "C" + col) ||
                       (!seen.add(number + "B" + row/3 + "," + col/3)))) {
                        return false;
                    }   
                }
            }
        }
        return true;
    }

chiragshahi (commented):
good job implementing multiple approaches
I have added my answer and code in the inline comment

chiragshahi:
good job implementing multiple approaches
I have added my answer and code in the inline comment

PR #56, opened 2020-05-26
PLease review

sbsreedh:


isValidSudoku.py
```diff
@@ -0,0 +1,26 @@
+# Time Complexity: O(n^2)
```
iivaishnavii, line 1: Its O(1) since you go through 81 elements

iivaishnavii:
Good job, Sonali.. Do look at the inline comments!

PR #62, opened 2020-05-27
Complete Desing

JacobStephen9999:


vaid sudoko.py
```diff
@@ -0,0 +1,21 @@
+# Time Complexity : O(N) where N is number elements in sudoko
```
iivaishnavii, line 1: Its O(1), since you just have 9*9 boxes to cover.

iivaishnavii:
Good job, Jacob! Keep the good work up!

PR #83, opened 2020-06-30
Add files via upload

trivedipratul:


PA_69.py
```diff
@@ -0,0 +1,95 @@
+#LC 36 - Valid Sudoku
+#Time Complexity - O(m*n)
+#Space Complexity - O(m*n)
+from collections import defaultdict
```
Lohith-Aswathappa, line 4: Since the size of the matrix is constant always, it'll be O(1)

PA_70.py
```diff
@@ -0,0 +1,33 @@
+#LC 515 - Find Largest Value in Each Tree Row
+#Time Complexity - O(n)
+#Space Complexity - O(n)
+# Definition for a binary tree node.
+# class TreeNode(object):
+#     def __init__(self, val=0, left=None, right=None):
```
Lohith-Aswathappa, line 6: Try the BFS approach as well.

Lohith-Aswathappa:
Good work, Pratul!
Review the inline comments.

PR #88, opened 2020-06-30
Completed both problems, please review.

samarthjj:


AbhiniveshP:
Nice work Samarth!!!
What change would there be in time and space complexities if you use append() and popleft() operations of deque, instead of appendleft() and pop() operations?

AbhiniveshP:
Nice work Samarth!!!
What change would there be in time and space complexities if you use append() and popleft() operations of deque, instead of appendleft() and pop() operations?

PR #92, opened 2020-07-01
maximum node in each level

mithi21:
please review it


AroopN:
Good work. Please also try solving "Valid Sudoku".

AroopN:
Good work. Please also try solving "Valid Sudoku".

PR #100, opened 2020-07-12
Abhijeet- PLease review

AbhijeetDhillon:


AroopN:
Good work, Abhijeet. The time complexity for valid Sudoku is also constant because the board size is always 9x9.

AroopN:
Good work, Abhijeet. The time complexity for valid Sudoku is also constant because the board size is always 9x9.

PR #258, opened 2021-03-25
completed competitive coding 5

alekhyagudi:


sudoko.py
```diff
@@ -0,0 +1,29 @@
+#Time complexity: O(M*N)
+#Space Complexity: O(M+N+O)
```
sibalnirbhay, line 2: Both would be O(1) as there are constant number of cells in a Sudoku grid,

sibalnirbhay (commented):
Read inline comments, otherwise looks good.

sibalnirbhay:
Read inline comments, otherwise looks good.

PR #296, opened 2021-06-28
Competitive-Coding-5 completed

krishnaattapu:
Competitive-Coding-5 completed


Rooppesh:
Good Job! In the Sudoku problem, since the size is always defined as 3x3, we can say that the TC and SC would be O(1)

Rooppesh:
Good Job! In the Sudoku problem, since the size is always defined as 3x3, we can say that the TC and SC would be O(1)

PR #538, opened 2022-08-23
files...please review

Sreelekhasamanu:


sudoku_validity.java
```diff
@@ -0,0 +1,55 @@
+// Time Complexity : O(n^2) n=9 iterating through each cell 81 cells foe rows, col, 3*3 grid
+// Space Complexity : O(n^2) n=9 considered as constant?
```
daredevilteja, line 2: No here you are using Hashsets of O(n**2) size in worst case

sudoku_validity.java
```diff
@@ -1,4 +1,4 @@
-// Time Complexity : O(n^2) n=9 iterating through each cell 81 cells foe rows, col, 3*3 grid
+// Time Complexity : O(n^2*n) n=9 iterating through each cell 81 cells for rows, col, 3*3 grid and through the set or boolean array for each search
```
daredevilteja, line 1: TC is O(n**2)

daredevilteja:
Good Work! Keep it up

PR #606, opened 2023-06-26
completed competative coding- 5

Sailaja-ASB:


problem2.py
```diff
@@ -0,0 +1,34 @@
+""" 
+Time Complexity : O(N) N= no.of nodes
+Space Complexity :O(m) size of queue
```
ishika2741, line 3: So what can be the maximum size of the queue in worst case? It will be all the leaf nodes. Which means SC = O(n) where n is the no. of nodes

ishika2741 (commented):
Well done! Keep it up.

ishika2741:
Well done! Keep it up.

PR #617, opened 2023-07-15
 maxValue in Each tree row problem solved

Erlybird:


poojaGit95:
Good job. Briefly explain the approach used

PR #654, opened 2023-11-03
Completed both solutions

merinjputhiyedath:


vyankateshmisar:
Good Job! In problem 2, the ultimate space and time complexity would be O(1).

vyankateshmisar:
Good Job! In problem 2, the ultimate space and time complexity would be O(1).

PR #673, opened 2024-01-22
Competitive-Coding-5

keerthana-vijaykumar:


Problem-1.cpp
```diff
@@ -0,0 +1,65 @@
+// Time Complexity : O(M*N)
```
ishika2741, line 1: Check time complexity plz

Problem-2.cpp
```diff
@@ -0,0 +1,51 @@
+// Time Complexity : O(M*N)
```
ishika2741, line 1: Kindly explain what m and n here is

ishika2741 (commented):
Good work!

ishika2741:
Good work!

PR #733, opened 2024-08-06
CC-5

ksurbhi:


Nandha2Tech (commented):
Great Work! TC  and SC for min cost tickets would be O(n).

Nandha2Tech:
Great Work! TC  and SC for min cost tickets would be O(n).

PR #779, opened 2025-01-26
Completed

NidhiPatel1410:


Erlybird:
Good job! Please mention the TC and SC of the 2nd problem

Erlybird:
Good job! Please mention the TC and SC of the 2nd problem

PR #820, opened 2025-02-24
Completed competitive-coding-5

sminannaware3:


Nandha2Tech (commented):
Great Work! Time complexity for brace-expansion would be exponential. k^n where k is the average size of each group and n is the number of groups.

Nandha2Tech:
Great Work! Time complexity for brace-expansion would be exponential. k^n where k is the average size of each group and n is the number of groups.

PR #833, opened 2025-03-03
done

rishabh-kr-jain:


234. Palindrome Linked List.py
```diff
@@ -0,0 +1 @@
+
```
sudhasr, line 1: Pls try this problem too!

sudhasr (commented):
Good Solution.

Please do try out the Palindrome Linked List problem too.

sudhasr:
Good Solution.

Please do try out the Palindrome Linked List problem too.

PR #874, opened 2025-06-16
Competitive Coding 5

HarshaReddyVardhan:
https://leetcode.com/problems/find-largest-value-in-each-tree-row/

https://leetcode.com/problems/valid-sudoku/


super30admin:
1. Correctness:
   - For "Find Largest Value in Each Tree Row", the solution correctly implements a depth-first search (DFS) approach to find the largest value in each level of the binary tree. The logic for updating the maximum value at each level is sound.
   - For "ValidSudoku", the solution correctly checks the validity of a Sudoku board by using sets to track seen numbers in rows, columns, and 3x3 sub-boxes. The logic for determining the sub-box index is correct.

2. Time Complexity:
   - For the tree problem, the time complexity is correctly stated as O(n), where n is the number of nodes in the tree, as each node is visited exactly once.
   - For the Sudoku problem, the time complexity is correctly stated as O(1) because the board size is fixed (9x9), and the operations per cell are constant time.

3. Space Complexity:
   - For the tree problem, the space complexity is correctly stated as O(h), where h is the height of the tree, due to the recursion stack. The output array is not included in the space complexity, which is appropriate.
   - For the Sudoku problem, the space complexity is correctly stated as O(1) because the number of sets and their sizes are fixed and do not scale with input size.

4. Code Quality:
   - The code is generally well-structured and readable. However, there are minor issues:
     - In the tree problem, the helper method could be made private since it's not intended to be called outside the class.
     - In the Sudoku problem, the variable name `bI` could be more descriptive (e.g., `boxIndex`).
     - Both solutions could benefit from brief comments explaining the high-level approach, especially for the Sudoku problem where the box index calculation might not be immediately obvious to all readers.

5. Efficiency:
   - Both solutions are efficient and follow standard approaches for their respective problems. No significant optimizations are needed.
   - For the tree problem, an iterative approach using BFS could also be considered, but the recursive DFS approach is equally valid and clean.

PR #907, opened 2025-07-23
Homework done

GuptaDivij:


super30admin:
Strengths:
- Both solutions are correct and handle edge cases appropriately.
- The binary tree solution efficiently uses BFS to traverse levels and find max values.

Areas for Improvement:
- For the Sudoku problem, consider pre-initializing HashSets outside loops to improve efficiency, as done in the reference solution.
- While the Sudoku solution is correct, the reference solution's approach of checking rows, columns, and boxes in a single pass is more elegant and slightly more efficient.
- The binary tree solution is well-optimized, but ensure that the queue is always properly managed (e.g., checking for null before polling is not necessary here but good practice in general).

PR #925, opened 2025-09-23
Add solutions for Largest Value in Tree Row and Valid Sudoku

ravithakur2k:
Added LargestValueTreeRow.py implementing BFS to find the largest value in each row of a binary tree, and ValidSudoku.py to validate a Sudoku board using sets for rows, columns, and 3x3 squares.


super30admin:
- Strengths:
  - The student has a clear understanding of the problem and has implemented an efficient solution using sets.
  - The code is clean, well-commented, and easy to follow.
  - The use of tuples to uniquely identify rows, columns, and sub-boxes is a good approach.
- Areas for Improvement:
  - The student could consider using separate sets for rows, columns, and sub-boxes (like in the reference solution) to make the code slightly more modular and easier to debug, though the current approach is also valid.
  - The comments could be expanded slightly to explain the logic behind the tuple usage in the sets, especially for beginners who might not immediately understand the approach.

PR #927, opened 2025-09-26
Competitive Coding 5

v-hemish:


super30admin:
1. Problem Mismatch: Ensure that the solution matches the problem description. The reference solution was for Sudoku validation, while your solution is for finding largest values in a binary tree. Always confirm the problem requirements before coding.
2. Testing: While the commented-out test case is useful, consider moving it to a separate test function or using a testing framework for better organization.
3. Edge Cases: The solution handles the empty tree case well (`if not root: return []`), which is good practice.

Overall, your solution is correct and efficient for the problem it addresses. Just be mindful of matching the solution to the correct problem in the future.

PR #941, opened 2026-01-10
Subhash1

subhash4061:


super30admin:
- The solution is correct and efficient, handling edge cases like an empty tree appropriately.
- The use of `deque` for BFS is a good choice for performance.
- The variable naming is clear (`maxi`, `queue`, `result`), making the code easy to understand.
- One minor improvement could be to avoid using `self.result` as an instance variable since it's not necessary for this problem. Instead, you could declare `result` as a local variable within the method.
- The comment at the top (`# TC-O(n) # SC-O(n)`) is helpful but could be expanded slightly to explain why these complexities are achieved (e.g., "BFS visits each node once, hence O(n) time and space").

PR #942, opened 2026-01-25
Compeitive-Coding-5 compelted

sujaygijre:


super30admin:
Strengths:
- The idea to use a single set to track row, column, and box constraints is innovative and can be efficient if implemented correctly.
- The code is readable and well-structured.

Areas for Improvement:
1. The main issue is that you are inserting strings for every cell, including empty ones. This is unnecessary and causes problems. You should only insert for non-empty cells. In your current code, you insert for every cell, but the duplicate check is only for non-empty cells. This means that for empty cells, you are adding strings that might later cause false duplicates. For example, if you have an empty cell at (0,0), you insert "r0.", "c0.", "b00.". Then if later you have a non-empty cell with value '5' at (0,0), you would try to insert "r05", "c05", "b005", but the set already has entries for the same row, column, and box from the empty cell? Actually, the empty cell entries are different because they have '.' so they won't conflict with digits. However, the problem is that you are inserting for empty cells, which is not needed and wastes space. More importantly, your duplicate check is done after the insertion for the current cell? Actually, no: you check before inserting? Let's look at the code:

    if (board[i][j]!='.' && ( us.contains(row) || us.contains(col) || us.contains(box) )) {
        return false;
    }
    us.insert(row);
    us.insert(col);
    us.insert(box);

So for a non-empty cell, you check if any of the strings already exist. But you have already inserted strings for previous cells, including empty ones. However, the strings for empty cells are different because they contain '.', so they won't match the strings for non-empty cells. So that part is safe. But wait: what about the same non-empty value in the same row? For example, if you have two '5's in the same row, then for the first '5' you insert "r05", and for the second '5' you check if "r05" exists, which it does, so you return false. That is correct.

But why is the solution failing? Actually, the student's solution might work for some cases, but there is a critical flaw: the insertion is done for every cell, including empty ones. However, the strings for empty cells are unique because they contain '.', so they won't conflict with non-empty cells. So why did the reference solution not do this? The reference solution only processes non-empty cells for row and column checks. In the student's solution, they are inserting strings for empty cells, which is unnecessary and increases the set size without any benefit.

But there is a more subtle issue: the box indexing formula. The student uses i/3 and j/3, which is correct. However, the reference solution uses a different formula for the box index in their loop? Actually, the reference solution has an error in the box index calculation in the provided reference solution? Wait, the reference solution provided in the problem has a mistake? Let me check the reference solution:

    int boxIdx = (i / 3 + j / 3 * 3);

This is correct because for i and j from 0 to 8, i/3 gives 0,1,2 and j/3 gives 0,1,2, so j/3*3 gives 0,3,6. Then i/3 + j/3*3 gives indices from 0 to 8. So that is correct.

But in the student's solution, they are using the same i and j for row, column, and box. So the box string is built with i/3 and j/3, which is correct.

However, the student's solution might fail because of the order of operations. Specifically, they are inserting the strings after the check. For a non-empty cell, they check if the strings are already in the set. If not, they insert. This is correct. But for empty cells, they insert without checking. This is not harmful because the strings for empty cells are unique and won't cause duplicates for non-empty cells. So why is the solution incorrect?

Wait, I see: the problem is that the student is using the same set for rows, columns, and boxes. The strings they generate for rows, columns, and boxes are unique because they have prefixes 'r', 'c', 'b'. So there is no conflict between a row string and a column string. So that is good.

But the student's solution should work correctly. However, it might fail in one edge: if the same value appears in the same row, column, or box, it will be detected. So why did I say it is incorrect? Let me test with Example 2.

Example 2 has a duplicate '8' in the top-left box. So for the first '8' at (0,0), it will insert:
    row: "r08"
    col: "c08"
    box: "b008"   (since i/3=0, j/3=0)

Then for the duplicate '8' at (0,0) in the modified board? Actually, in Example 2, the duplicate is in the top-left box? The top-left box has two '8's: one at (0,0) and one at (2,3)? Wait, no: in Example 2, the change is that the top-left corner (0,0) is changed from '5' to '8'. So the board has two '8's in the top-left box: at (0,0) and at (2,3)? Actually, the top-left box includes rows 0-2 and columns 0-2. The duplicate is at (0,0) and (2,3) is not in the same box. Wait, let's look at Example 2 board:

    ["8","3",".",".","7",".",".",".","."]
    ,["6",".",".","1","9","5",".",".","."]
    ,[".","9","8",".",".",".",".","6","."]

The duplicate '8's are at (0,0) and (2,2)? because (2,2) is '8'. So for (2,2), the box string is "b028". But the set already has "b008" from (0,0). So when processing (2,2), it will check for "b028" and see that it is not in the set? But wait, the box index for (2,2) should be i=2, j=2 -> i/3=0, j/3=0? So the box string is "b008" for both? So the set already has "b008" from (0,0), so when processing (2,2), it will try to insert "b008" again and the check will fail? So it should return false.

So the student's solution should work for Example 2.

But why did I think it was incorrect? Because of the empty cells. However, the empty cells are not causing any harm. So the solution might actually be correct.

Wait, but the student's solution has a mistake in the box string: they use i/3 and j/3, which is correct. But they are using integer division, which in C++ truncates towards zero, so for i=0,1,2 -> i/3=0, i=3,4,5->1, etc. So it is correct.

So the student's solution is correct? Then why did the reference solution use a different approach? The reference solution uses separate sets for each row, each column, and each box. The student's solution uses one set for all, but with unique strings.

However, there is one more

PR #947, opened 2026-01-27
Competitive-Coding-5 Done

Sreeja-99:


super30admin:
Your solution for the Sudoku problem is correct and efficient. Well done! Here are a few minor suggestions to improve code quality:

1. Avoid using raw types. Instead of `Set r = row.get(i);`, use `Set<Integer> r = row.get(i);`. This ensures type safety and makes the code more robust.

2. The variable `req` for the box index is correctly calculated, but you might want to explain the formula in a comment for clarity. For example: `// Box index = (row_index/3)*3 + column_index/3`.

3. The space complexity comment says "3*O(9*9)", which is misleading. Since each set can hold at most 9 elements and there are 9 sets per type, the total space is 3 * 9 * 9 = 243 integers at worst, which is constant. So it's O(1). You should note that.

4. In the Sudoku solution, you have a variable `m` and `n` for the dimensions, but since the grid is always 9x9, you can simply use 9. However, using `m` and `n` is good for generality.

Overall, your solution is excellent. Keep up the good work!

PR #949, opened 2026-02-07
Submitting Q1

akshay4121:


super30admin:
It seems there has been a mix-up in the solutions. The code you submitted is for "Largest Values in Each Tree Row", but the problem you need to solve is "Valid Sudoku". Please ensure you are working on the correct problem.

For the Sudoku problem, you need to validate a 9x9 Sudoku board according to the given rules. You should check for duplicates in each row, each column, and each 3x3 sub-box. A common approach is to use sets or arrays to track the digits seen in each row, column, and box. 

Here are some tips for solving the Valid Sudoku problem:
- Use three arrays of sets (or boolean arrays) for rows, columns, and boxes.
- Iterate through each cell in the board. For each non-empty cell, check if the digit has been seen in the current row, column, or box. If it has, return false immediately.
- The box index can be calculated as (row / 3) * 3 + (col / 3).
- Remember that the problem only requires validating filled cells; empty cells ('.') should be skipped.

Please resubmit your solution for the correct problem. If you need help with the Sudoku problem, let me know, and I can provide guidance.

super30admin:
It seems there has been a mix-up in the problems. You have submitted a solution for "Largest Value in Each Tree Row" instead of "Valid Sudoku". For the Sudoku problem, you need to validate a 9x9 grid according to the rules of Sudoku. 

To solve the Valid Sudoku problem:
- You need to check each row for duplicates.
- Check each column for duplicates.
- Check each 3x3 sub-box for duplicates.
- You can use sets or arrays to track seen digits in rows, columns, and boxes.

For example, you can create arrays of sets for rows, columns, and boxes. Then iterate through each cell, and for each non-empty cell, check if the digit exists in the corresponding row set, column set, and box set. If it does, return false; otherwise, add it to the sets.

Here is a corrected approach for Valid Sudoku:
- Initialize 9 sets for rows, 9 for columns, and 9 for boxes.
- For each cell (i, j):
  - If the cell is not empty:
    - Check if the digit is in rowSet[i]. If yes, return false; else add it.
    - Check if the digit is in colSet[j]. If yes, return false; else add it.
    - Calculate the box index: boxIndex = (i/3)*3 + j/3. Check if the digit is in boxSet[boxIndex]. If yes, return false; else add it.
- If all checks pass, return true.

Please ensure you are solving the correct problem in the future.

PR #953, opened 2026-02-27
Completed Competitive Programming 5

eyeamkd:


super30admin:
It seems there might be a misunderstanding. The problem is to validate a Sudoku board, not to solve it. Your solution is for solving the Sudoku, which is a different problem. 

For the validation problem, you need to check:
1. Each row has no duplicates (except '.').
2. Each column has no duplicates.
3. Each 3x3 sub-box has no duplicates.

You can do this by using sets for each row, column, and box. Since the grid is fixed at 9x9, you can create 9 sets for rows, 9 for columns, and 9 for boxes.

Here's a suggested approach:
- Initialize 9 sets for rows, 9 for columns, and 9 for boxes.
- Iterate over each cell in the grid.
- If the cell is not '.', check if the value exists in the corresponding row set, column set, or box set. If it does, return false.
- Otherwise, add the value to the respective sets.

The box index can be calculated as `box_index = (i // 3) * 3 + j // 3`.

Your current solution for solving Sudoku is not required for this problem. Also, the code for the binary tree problem is unrelated.

Please revise your solution to address the correct problem. Remember that for validation, you don't need to solve the board, only check its current state.

PR #955, opened 2026-03-08
ValidSudoku and BFS

viveksharma180:


super30admin:
### Interview Problem: Valid Sudoku (validSudoku.py)
Strengths:
- You have written clear code with comments, which is good practice.
- You considered the problem and attempted to solve it by checking rows, columns, and blocks.

Areas for improvement:
- The approach of scanning the row and column for each cell is inefficient and leads to redundant operations. Instead, you should use data structures (like sets or arrays) to keep track of numbers seen in each row, column, and block. This allows you to check each cell only once.
- The 3x3 block check is incorrect: you should not exclude the current cell by `r != i and c != j` because you are iterating over the entire block. Actually, you should compare all cells in the block except the current one? But note: the current cell is included in the block. So when you check the block, you are comparing the current cell with every cell in the block, including itself. Therefore, you must skip the current cell. However, your condition `r != i and c != j` is not sufficient: if there is a duplicate in the same block at (r, c) that is not the same row or column as (i,j), it would be caught. But if the duplicate is in the same row but different column? Then `r==i` but `c != j`, so it would be skipped? Actually, no: because your condition requires both `r != i` AND `c != j`. So if there is a duplicate in the same row but different column, `r==i` so the condition fails and it is not considered. This means you might miss duplicates in the same row of the block. Similarly for same column. Therefore, the condition should be `if not (r == i and c == j)`.

But actually, the entire method of checking the block by iterating for every cell is inefficient. It would be better to use a set per block.

- The problem requires that each row, column, and block has no duplicates. Your method checks for duplicates in the row by scanning left and right from the current cell. However, this does not ensure the entire row is unique. For example, if there is a duplicate in the same row but both to the left and right of the current cell, your checks (left and right) would catch it. But actually, when you check the left part for each cell, you are effectively checking the entire row? Let's see: for a given row, you check for each cell all the cells to the left. So by the time you get to the rightmost cell, you have compared it with all left cells. Similarly, when you check the right part for the leftmost cell, you check all right cells. So the row is entirely checked? Actually, no: because for a cell at position j, you check all left (0 to j-1) and all right (j+1 to 8). So indeed, the entire row is checked for duplicates. Similarly for the column. So the row and column checks should be correct. However, the issue is that the same row is checked multiple times (for each cell in the row) which is inefficient.

But wait: the problem says only filled cells need to be validated. So if there are duplicates, we only care about the filled cells. Your method checks every filled cell against the rest of the row and column. This should work for row and column duplicates.

However, the block check is flawed as explained above.

Another issue: the block check is done for every cell. So the same block is checked multiple times (9 times). This is inefficient.

But the major flaw is in the block check condition. Let me correct it: you should check all cells in the block except the current cell. So the condition should be:
if (r != i or c != j) and board[i][j] == board[r][c]:
    return False

But note: this would also catch duplicates that are in the same row or column? Yes, but that is correct because the block must be unique.

However, there is a better way: use sets for blocks.

Recommendation: Use the reference solution approach. Maintain 9 sets for rows, 9 for columns, and 9 for blocks. For each cell (i, j), if it is filled, check if the value exists in the set for row i, column j, or box (i//3, j//3). If it does, return false. Otherwise, add it to the sets.

This is more efficient and straightforward.

VERDICT: NEEDS_IMPROVEMENT

---

### Interview Problem : Greatest Value in Each Row (largestInEachLevel.py)
Your solutions are excellent! You have provided two different approaches (BFS and DFS) that both correctly solve the problem. Here are some strengths and minor suggestions:

Strengths:
- You have correctly analyzed the time and space complexity for both solutions.
- The code is clean, well-commented, and follows good naming conventions.
- You have handled edge cases (like root being null) appropriately.
- Using a deque for BFS is efficient in Python.

Areas for Improvement:
- For the DFS solution, consider that using an instance variable (self.result) might not be the most flexible approach if the method is called multiple times (since self.result would retain state between calls). However, in the context of Leetcode, where the solution is run once per test case, this is acceptable. Alternatively, you could avoid using an instance variable by passing the result list as a parameter or using a helper function that returns the result. For example:
  ```python
  def largestValues(self, root: Optional[TreeNode]) -> List[int]:
        result = []
        self.dfs(root, 0, result)
        return result

  def dfs(self, node, depth, result):
        if not node:
            return
        if depth == len(result):
            result.append(node.val)
        else:
            result[depth] = max(result[depth], node.val)
        self.dfs(node.left, depth+1, result)
        self.dfs(node.right, depth+1, result)
  ```
  This avoids state in the instance and makes the function more self-contained.

- In the DFS solution, you have checks for `if root.left is not None` and similarly for right. You can simplify these to `if node.left:` and `if node.right:` for brevity, which is idiomatic in Python.

- You might want to add a brief comment in the DFS solution explaining how it works (e.g., "Using DFS preorder traversal, we update the maximum value for each depth as we traverse").

Overall, your solutions are robust and efficient. Great job!

VERDICT: PASS

PR #958, opened 2026-03-11
Complete Competitive-Coding-5

PrakarshKamal:


super30admin:
### Interview Problem: Valid Sudoku (ValidSudoku.java)
Strengths:
- You recognized that using a HashSet to track seen digits is a good approach.
- You attempted to avoid three separate data structures by using one set with unique strings.

Areas for Improvement:
1. **Indexing**: Remember that Java arrays are 0-indexed. Your loops should run from 0 to 8 (inclusive). The inner loop condition `j <= 0` is incorrect and should be `j < 9` or `j <= 8`.
2. **Handling Empty Cells**: You must skip cells that contain '.' because only filled digits need validation.
3. **String Construction**: Your strings for row, column, and block need to be unique and correctly formatted. For example:
   - For row: `num + " in row " + i` is correct, but note that `i` should be the row index.
   - For column: It should be `num + " in column " + j` (not "at row j").
   - For block: You need to compute a unique block index. The common way is `3*(i/3) + (j/3)` (integer division) which gives a number from 0 to 8. Then you can use `num + " in block " + blockIndex`.
4. **Duplicate Checks**: You are checking for `col` twice instead of `block`. Also, you should check each string only if the cell is not empty.
5. **Efficiency**: Your approach is efficient in terms of time and space, but you need to correct the implementation.

Suggested corrected code (based on your approach):
```java
class Solution {
        Set<String> seen = new HashSet<>();
        for (int i = 0; i < 9; i++) {
            for (int j = 0; j < 9; j++) {
                char num = board[i][j];
                if (num != '.') {
                    String row = num + " in row " + i;
                    String col = num + " in column " + j;
                    int blockIndex = (i/3)*3 + j/3; // This gives a unique index from 0 to 8 for each 3x3 block.
                    String block = num + " in block " + blockIndex;
                    if (seen.contains(row) || seen.contains(col) || seen.contains(block)) {
                        return false;
                    }
                    seen.add(row);
                    seen.add(col);
                    seen.add(block);
                }
            }
        }
        return true;
    }
}
```

VERDICT: NEEDS_IMPROVEMENT

---

### Interview Problem : Greatest Value in Each Row (LargestValueInEachRow.java)
The student's solution is correct and efficient. The code is clean and follows good practices. However, note that in a competitive programming or interview setting, it is important to consider edge cases. For example, if the tree is very large, but the solution handles that with O(n) complexity. Also, the student should be aware that the `TreeNode` class is typically provided in online judges, so they might not need to define it themselves.

One minor improvement: Instead of using `int max = Integer.MIN_VALUE;`, if the level has no nodes (which is not possible in this code because we are processing a non-empty queue), but to be absolutely safe, we could consider that the level might be empty. However, since we are starting with root non-null and only adding non-null children, every level we process has at least one node. So it is safe.

Another minor point: The student uses `q.offer(root)` and `q.poll()` which is good because `offer` and `poll` are preferred in Java for queues (as they return false/null instead of throwing exceptions). But note that `LinkedList` in Java implements `Queue` and these methods are correct.

VERDICT: PASS

PR #960, opened 2026-03-12
Competitive-Coding-5

pranjay01:


super30admin:
### Interview Problem: Valid Sudoku (Interview5.py)
It seems there has been a mix-up in your submission. You have provided a solution for "Find Largest Value in Each Tree Row" (a binary tree problem) instead of the "Valid Sudoku" problem. Please review the problem statement and submit a solution that addresses the Sudoku validation rules:

- Check each row for duplicates (digits 1-9 without repetition).
- Check each column for duplicates.
- Check each 3x3 sub-box for duplicates.

For the Sudoku problem, you can use sets or arrays to track seen digits in rows, columns, and boxes. The reference solution uses three arrays of sets (for rows, columns, and boxes) and iterates through the 9x9 grid once to check all conditions.

VERDICT: NEEDS_IMPROVEMENT

---

### Interview Problem : Greatest Value in Each Row (Interview5_2.py)
Your solution shows you're thinking about the problem, but has several issues:
1. The core approach is flawed - Sudoku validation should check for duplicates, not track "remaining" numbers
2. You should use boolean arrays or sets to track seen numbers in each row, column, and box
3. There are syntax errors (missing colons in for loops)
4. Consider using three 9x9 boolean arrays to track which numbers have appeared in each row, column, and box
5. The problem requires checking existing numbers only - ignore "." entries but check that all numbers 1-9 appear at most once per row/column/box

VERDICT: NEEDS_IMPROVEMENT

PR #961, opened 2026-03-15
Add implementation for leetcode problems 36, 515

rishigoswamy:


super30admin:
### Interview Problem: Valid Sudoku (leetcode_36.py)
Your solution is correct and efficiently checks the Sudoku board. You have successfully implemented the required checks for rows, columns, and sub-boxes. Here are some strengths and areas for improvement:

Strengths:
- You correctly identified the need to check rows, columns, and boxes separately.
- The code is well-structured with clear comments explaining the approach.
- You used sets to track duplicates, which is an efficient way to check for repetitions.

Areas for Improvement:
1. Efficiency: Your solution uses two passes (one for rows/columns and one for boxes), which results in 162 operations for a 9x9 grid. The reference solution does it in one pass (81 operations). While both are O(1), you can optimize by checking rows, columns, and boxes in a single pass. This would reduce the number of iterations and make the code more efficient.

2. Code Readability: The row and column checks are combined in the same loop, which might be slightly confusing. Consider separating them for clarity, or at least add comments to explain each part.

3

VERDICT: NEEDS_IMPROVEMENT

---

### Interview Problem : Greatest Value in Each Row (leetcode_515.py)
Strengths:
- The code is well-commented and the student has clearly explained their approach.
- The student recognized that BFS is suitable for level-order traversal.

Areas for improvement:
1. Correctness: The current implementation has a bug in handling the last level. For example, with a single node tree, the output will have an extra -inf. You should test with a small example (like root = [1]) to see the issue. The standard way to avoid this is to not rely on a delimiter for level boundaries. Instead, you can use a for-loop that processes exactly the number of nodes in the current level (which is the size of the queue at the start of the level). This is what the reference solution does.

2. Efficiency: Avoid using `list.pop(0)` because it is O(n). Instead, use `collections.deque` which allows O(1) appends and pops from both ends. This will improve the time complexity to O(n).

3. Simplification: The delimiter method is more complex than necessary. The reference solution shows a simpler approach: at the beginning of each level, the queue contains all nodes of that level. You can get the size of the queue (number of nodes at the current level) and then process exactly that many nodes, updating the maximum for the level. Then you add the children for the next level. This avoids the need for a delimiter and reduces the chance of errors.

4. Edge Cases: Your code handles the case when root is None correctly. However, the bug with the last level remains.

Revised Approach:
- Use a deque for the queue (for efficient pops).
- While the queue is not empty:
   - Get the current level size (number of nodes in the queue at the start of the level).
   - Initialize max_val for the level to -inf.
   - For each node in the current level (i in range(size)):
        Pop the node from the left of the deque.
        Update max_val = max(max_val, node.val)
        If the node has left child, append it to the deque.
        If the node has right child, append it to the deque.
   - Append max_val to the result list.

This approach is cleaner and more efficient.

VERDICT: NEEDS_IMPROVEMENT

PR #962, opened 2026-03-24
Find largest value in each tree node solution added

megharaykar:


super30admin:
### Interview Problem: Valid Sudoku (findLargestInEachRow.py)
It appears there has been a mix-up in the problem you were solving. The problem you solved is "Find Largest Value in Each Tree Row", but the assigned problem is "Valid Sudoku". Please ensure you are working on the correct problem.

For the Sudoku problem, you need to validate a 9x9 grid according to the three rules: no duplicates in rows, columns, and 3x3 sub-boxes. You should use sets or arrays to track seen numbers in each row, column, and box. The reference solution provided uses three arrays of sets (for rows, columns, and boxes) and checks each cell in a nested loop.

Here are some tips for the Sudoku problem:
- Use sets for each row, column, and box to check for duplicates.
- Remember that the board is 9x9, so you can use fixed-size arrays.
- For the box index, you can calculate it as `(i // 3) * 3 + j // 3` (in Python) to map each cell to one of the 9 boxes.

You should start over with the correct problem. Your tree solution is good for that problem, but it doesn't apply here.

VERDICT: NEEDS_IMPROVEMENT

---

### Interview Problem : Greatest Value in Each Row
Strengths:
- The solution uses BFS which is appropriate for level-order traversal.
- The code is clear and well-structured, with comments indicating time and space complexity.
- The base case for an empty tree is handled correctly.
- The use of `deque` from `collections` for efficient popping from the front is good practice.

Areas for Improvement:
1. The code does not import `deque`. While it might be assumed that the student intended to import it, the code as written would raise a `NameError` because `deque` is not defined. You should add `from collections import deque` at the top.
2. The variable name `qroot` is a bit confusing. It might be better to use `node` or `current` to indicate that it represents the current node being processed.
3. The initial value for `largest` is set to `float('-inf')`, which is appropriate for Python. However, note that the problem constraints state that node values can be as low as `-2^31`, which is within the range of `float('-inf')` but since we are dealing with integers, it is acceptable. Alternatively, you could use `-10**18` or similar to avoid floating point issues, but `float('-inf')` is standard and safe in this context because we are only comparing integers.
4. The code does not have any major issues, but it's always good to ensure that all necessary imports are included.

VERDICT: PASS

PR #963, opened 2026-04-14
Complete Competitive-Coding-5

paridhimalviya:
Is liked list palindrome.
Height balanced binary tree


super30admin:
### Interview Problem: Valid Sudoku (HeightBalancedBinaryTree.swift)
It appears there has been a mix-up in the solution file. The problem you were asked to solve is "Valid Sudoku", but the code you provided is for "Height-Balanced Binary Tree". Please ensure you are working on the correct problem.

For the Valid Sudoku problem, you need to check three conditions: no duplicates in any row, no duplicates in any column, and no duplicates in any 3x3 sub-box. An efficient approach is to use sets to track seen digits in each row, column, and box. Since the grid is fixed at 9x9, you can initialize 9 sets for rows, 9 for columns, and 9 for boxes.

Here's a brief outline of how to approach the Sudoku problem:
1. Initialize sets for rows, columns, and boxes (each of size 9).
2. Iterate through each cell in the 9x9 grid.
3. For each non-empty cell, check if the digit is already in the corresponding row set, column set, or box set. If it is, return false.
4. Otherwise, add the digit to all three sets.
5. If you finish the iteration without duplicates, return true.

You can compute the box index for a cell (i, j) as (i/3)*3 + j/3 (using integer division).

Please review the problem statement again and implement the solution for Valid Sudoku. If you have any questions about the problem or the approach, feel free to ask.

VERDICT: NEEDS_IMPROVEMENT

---

### Interview Problem : Greatest Value in Each Row (IsLinkedListPalindrome.swift)
It seems there has been a mix-up in the problem you were solving. The problem you submitted code for is "Is Linked List Palindrome", but the problem you were supposed to solve is "Greatest Value in Each Row", which involves a binary tree, not a linked list. 

For the correct problem, you need to traverse the binary tree level by level (using BFS) and find the maximum value in each level. Your solution should involve a queue to perform level-order traversal. Here are some steps to consider:
1. If the root is null, return an empty array.
2. Use a queue to store nodes at each level.
3. For each level, initialize a variable to store the maximum value (start with INT_MIN).
4. Process all nodes in the current level, updating the maximum value and enqueuing their children.
5. After processing the level, add the maximum value to the result array.

Your current code does not address this problem at all. Please review the problem statement again and implement the solution for the binary tree problem.

VERDICT: NEEDS_IMPROVEMENT

